PR #16, opened 2016-05-11
Fix wrong counts in query details for queries on system objects

### LINKED ISSUE #15: Number of executions can be incorrect in query detail view
PoWA Web 3.0.2

If the same query is executed in multiple databases, the query detail view will report the wrong # of executions (and runtime). For example, if 4 databases have the same query, the # of executions will be 4 times the real value.

From what I can see, this is because in query.py's `class QueryDetail`, there's a join with powa_statements but the join is only on queryid. Since queryid is the same for the same query in all databases, it will return a row for each database (dbid) and those will be grouped and summed giving wrong results.

Changing the code to also join on dbid seems to correct the problem. The fix I did is champeric/powa-web@c490b3130b, but I'm not familiar with SQLAlchemy nor PoWA's code in general so I don't know what else it could affect.

Thanks

champeric:
We need to join on powa_statements with both queryid and dbid.

For queries on system objects, the queryid will be the same across databases.

Should fix #15 .

### DIFF
--- a/powa/query.py
+++ b/powa/query.py
@@ -5,7 +5,7 @@ Dashboard for the by-query page.
 from tornado.web import HTTPError
 from sqlalchemy.sql import (
     bindparam, text, column, select,
-    case, cast,
+    case, cast, and_,
     func, extract)
 from sqlalchemy.sql.functions import sum
 from sqlalchemy.types import Numeric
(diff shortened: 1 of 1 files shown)


rjuju:
Thanks!

PR #21, opened 2017-02-14
Fixe explain query with like statement

dblugeon:
Hello Dalibo team
This pull request fixe "TypeError: 'dict' object does not support indexing" error by adding call text function from sqlalchemy.

This error occurs when the powa-web application explains a query with like statement '%'.

### DIFF
--- a/powa/query.py
+++ b/powa/query.py
@@ -239,7 +239,9 @@ class QueryExplains(ContentWidget):
                 query = format_jumbled_query(row['query'], vals['constants'])
                 plan = "N/A"
                 try:
-                    result = self.execute("EXPLAIN %s" % query,
+                    from sqlalchemy import text
+                    sqlQuery = text("EXPLAIN %s" % query)
+                    result = self.execute(sqlQuery,
                                           database=database)
                     plan = "\n".join(v[0] for v in result)
                 except:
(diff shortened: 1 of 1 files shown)


rjuju:
Hello.

Indeed, thanks a lot for the patch!  Sorry for the delay.

Just some nitpicking:

- could you move to import at the top of the file with the other one
- can you confirm that text() is available since at least sqlalchemy 0.8.0

Thanks in advance.

dblugeon:
Hello @rjuju 
Sorry for the delay. ^^
I fix the import text. After checking, the sqlalchemy.text function is an alias to sqlalchemy.sql.text.

I tested with an virtualenv with sqlalchemy==0.8 . It's ok.

Sorry for the commits,I failed my rebasing. Do you want I remake my pr ?

rjuju:
Thanks for checking!, and sorry for the delay.

I personnally prefer clean git history. Maybe you can just rebase locally and push force, the PR should be automatically updated. Otherwise I can handle it directly and commit it with you as the author.

dblugeon:
Thanks for your suggestion about push force. It worked!

The pr history is more cleaner.

rjuju:
Ah good news:)

Thanks a lot for the fix!

PR #28, opened 2017-05-16
Fix wrong calculation of microseconds difference

champeric:
The isAfter function seems to only compare up to milliseconds so we have
to manually compare for more precision if m1 and m2 seem equal.

The calculation for microsecDiff was inverted.

Should fix dalibo/powa#88 .


rjuju:
You're absolutely right. Thank you very much!

PR #32, opened 2017-06-15
Use grunt-sass instead of grunt-contrib-sass

pgiraud:
As opposed to `grunt-contrib-sass`, `grunt-sass` doesn't require Ruby nor Sass to be installed on the system. It relies on `node-sass` which makes installation a bit easier.

Note: modifications on the built `.min.css` are due to precision in rounded values or formatting.


pgiraud:
This has been integrated. See 9fd53b5e84cc4ebdef94821bd777db1da045e847. Closing.

PR #34, opened 2017-07-11
Typo (double negation)

pgiraud:


rjuju:
Thanks for noticing the double negation.  You're fix changed the meaning which was correct. I pushed https://github.com/dalibo/powa-web/commit/16b1f21c2e722f45816ca65ee8176b325199f6eb to fix this unwisely chosen wording.

Thanks!

PR #38, opened 2017-09-29
Show a sorted list of servers on login screen

meikomeik:
We are starting to use powa with >1000 servers and the unsorted list of hostnames is an issue for us.
We would be really happy if this one liner could get merged.

### DIFF
--- a/powa/templates/login.html
+++ b/powa/templates/login.html
@@ -8,7 +8,7 @@
         {% apply field(name="username", label=_("Username")) %}{% end %}
         {% apply field(name="password", label=_("Password"), type="password") %}{% end %}
         {% apply field(name="server", label=_("Server"), tag="select") %}
-          {% for opt in servers().keys() %}
+          {% for opt in sorted(servers().keys()) %}
           <option value="{{opt}}">{{opt}}</option>
           {% end %}
         {% end %}


rjuju:
That's quite a huge number of servers.  This PR is ok so I'll merge it now, thanks a lot.  Perhaps we could also implement an autocompletion with filtering?

meikomeik:
I like that idea though I can not promise that I'm capable enough of doing this in an elegant way. 
Will look into it.

rjuju:
Actually I was suggesting that I do it, if that's of any interest for you :)  If you want to implement it I'll gladly accept a PR though, it's up to you.

meikomeik:
I haven't touched Java script for some time so I'd be happy if you could do it. 
I would probably also be capable but it would take me much more time.

rjuju:
Ok! I'll try to do it soon

pgiraud:
@meikomeik In order to make sure that we find the adequate user interaction to solve this "long list of servers" issue, it would be appreciated if you could provide us some answers to the following questions :
 - what do you server names look like? Please give 2 or 3 different examples.
 - when you are looking for a server, are you searching for some text at the begining of the name? Or would a more complex matching pattern help?
My opinion is that we don't need to add complexity if it's not worth.

pgiraud:
@meikomeik Additional question: do you often switch from one server to an other? If so, it means that you first need to logout to do so, right?

PR #45, opened 2017-11-21
Clear off table

pgiraud:
The goal of this PR is to improve readability of the data in the grid tables.

Before
screenshot from 2017-11-24 16-34-45
After
screenshot from 2017-11-24 16-35-43


powa/static/js/powa/utils/duration.js
```diff
@@ -19,10 +19,10 @@ define(['moment'], function(moment){
         microseconds: 'µs',
         delimiter: ' '
     };
-    moment.fn.preciseDiff = function(d2) {
-        return moment.preciseDiff(this, d2);
+    moment.fn.preciseDiff = function(d2, imprecise) {
```
rjuju: Probably at some point we'll want to make this configurable, so I think it'd be better to have this parameter given as a precision, with -1 (or not specified) meaning no rounding, and otherwise round at the given precision.  What do you think?

rjuju:
Thanks a lot for this PR, grids are way better after applying the modifications!

Everything is fine, so I'll merge it soon.  I'm just waiting on your opinion on previous comment.

powa/static/js/powa/utils/duration.js
```diff
@@ -108,6 +108,10 @@ define(['moment'], function(moment){
         if(result.length == 0){
             return "0";
         }
+        // Keep only the first 2 results, more it too precise
```
pgiraud: s/it/is/

marco44:
Yes, definitely to be merged :)

powa/static/js/powa/utils/duration.js
pgiraud: I understand your point. I just wanted to keep it simple.
Need precision? -> set it to false or don't use it
Don't need precision? -> set it to true, we choose the precision for you.

Choosing the precision could make sense though.

rjuju: That works for me.  It won't be hard to change later anyway.

rjuju:
Merged, thanks a lot!

PR #46, opened 2017-11-21
Update compiled css

pgiraud:
This needs update because the recent change of sass compiler introduces
changes in floating number rounding

Follow up for 9fd53b5e84


rjuju:
Thanks!

PR #49, opened 2018-06-15
Set grunt-sass version

pgiraud:
- updates grunt-sass to the latest version along with generated css modifications,
 - prevents unintended grunt-sass updates in the future.


rjuju:
Thanks!

PR #51, opened 2018-06-15
Fix broken usage of label for fields

pgiraud:
We most probably don't want the `username` as id for an input field.


rjuju:
Thanks!

PR #52, opened 2018-06-15
Better layout for login page

pgiraud:
Only costmetics.

Before:
0 0 0 0_8888_login__next 2f laptop with mdpi screen 1

After:
0 0 0 0_8888_login__next 2f laptop with mdpi screen


rjuju:
Thanks!

PR #54, opened 2018-06-15
Query detail

pgiraud:
Layout improvements for the query detail page.
 - Smaller font for the SQL block,
 - Max height set with scrollbar if required,
 - Below info centered.

Before:
0 0 0 0_8888_database_powa_query_421932566_overview_ laptop with mdpi screen 1

After:
0 0 0 0_8888_database_powa_query_421932566_overview_ laptop with mdpi screen


rjuju:
Thanks!

PR #57, opened 2018-06-18
Remove invisible bug icon

### LINKED ISSUE #35: icons don't show up
Icons are supposed to be displayed (in the footer for example) but they aren't because fontawesome is not loaded.

pgiraud:
Fixes #35

We don't use fontawesome and unfortunately there's no bug icon in foundation icons set.
Let's remove it.


rjuju:
Yes that's unfortunate.  Fortunately there's no bug in powa-web :)

PR #67, opened 2018-12-18
Reapply search filter when new data is received

### LINKED ISSUE #66: filter is lost when changing the time range
On the query list.

Filter on a string, then change the time range, filter is still displayed, but not applied anymore

pgiraud:
Fixes #66


rjuju:
actually I don't understand the original issue.  @marco44 (or @pgiraud if you got it) could you provide more details on what's the problem is, and how to trigger it?

pgiraud:
Here's the steps to reproduce the issue (on master):
 - open any page in which there's a grid (either "all databases" or "/database/your_db/overview"),
 - enter a value in the search field,
 - make sure the data is filtered adequately (ie. that you have a lower number of rows in the grid),
 - change the date range.

On master, after the date range is changed, even though the search field still displays the text previously entered by user, the data in the grid is not filtered anymore.

With the current PR, the search filter is applied once new data is received.

But you were right to mention browser caching issue. Because what I forgot is to compile javascript. I'm running powa in debug mode.

PR update to come.

rjuju:
Perfect!  Thanks a lot and sorry for  the delay.

PR #68, opened 2018-12-21
Multiuser fixes

rdunklau:
Small fixes regarding multiple user and performance problems with quals.

### DIFF
--- a/powa/qual.py
+++ b/powa/qual.py
@@ -104,7 +104,8 @@ class QualDetail(ContentWidget):
             FROM powa_qualstats_quals
             JOIN powa_statements USING (queryid)
             WHERE qualid = :qual
-            LIMIT 5"""), {"qual": qual})}
+            LIMIT 5"""),
+            params={"qual": qual})}
         for qual in quals:
             if qual['is_my_query']:
                 my_qual = resolve_quals(remote_conn, [qual])[0]
--- a/powa/qual.py
+++ b/powa/qual.py
@@ -4,7 +4,7 @@ Dashboard for the qual page
 from sqlalchemy.sql import text, bindparam
 from tornado.web import HTTPError
 from powa.dashboards import (
-    Dashboard, Graph,
+    Dashboard, Graph, Grid,
     MetricGroupDef, MetricDef,
     DashboardPage, ContentWidget)
 from powa.sql import qual_constants, resolve_quals
@@ -99,13 +99,6 @@ class QualDetail(ContentWidget):
         my_qual = None
         other_queries = {}
 
-        other_queries = {r['queryid']: r['query'] for r in self.execute(text("""
-            SELECT DISTINCT queryid, query
-            FROM powa_qualstats_quals
-            JOIN powa_statements USING (queryid)
-            WHERE qualid = :qual
-            LIMIT 5"""),
-            params={"qual": qual})}
         for qual in quals:
             if qual['is_my_query']:
                 my_qual = resolve_quals(remote_conn, [qual])[0]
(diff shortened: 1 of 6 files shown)


rjuju:
@rdunklau last commit should fix the already existing issue I was talking about.

powa/qual.py
```diff
@@ -72,23 +78,28 @@ def get(self, database, query, qual):
...
+                    "queryids": [query],
                     "qualid": qual}))
         my_qual = None
         other_queries = {}
+        other_queries = {r['queryid']: r['query'] for r in self.execute(text("""
+            SELECT DISTINCT queryid, query
+            FROM powa_qualstats_quals
+            JOIN powa_statements USING (queryid)
+            WHERE qualid = :qual
+            LIMIT 5"""), {"qual": qual})}
```
rjuju: this **5** seems a little bit arbitrary.  As far as I can tell it shouldn't raise performance issues to try to retrieve more queries involving the same qual, and it seems like an interesting information to have, especially if there are more than 5 queries having the same qual.  What about raising it quite a bit (50 or so?)

Also, shouldn't the query exclude the current queryid?

powa/qual.py
```diff
@@ -72,23 +78,28 @@ def get(self, database, query, qual):
...
+            .column((stmt.c.queryid == bindparam("query")).label("is_my_query")))
         quals = list(self.execute(
             stmt,
             params={"query": query,
                     "from": self.get_argument("from"),
                     "to": self.get_argument("to"),
+                    "queryids": [query],
                     "qualid": qual}))
         my_qual = None
         other_queries = {}
```
rjuju: is it still needed?

powa/qual.py
rdunklau: Performance wise, it's not a problem but the current UI makes you scroll for kilometers before seeing any other information.
Those queries should probably moved to a separate grid widget, but in the meantime that seemed like the best option.

rjuju: What about adding an overflow attribute to the div?  It'll also fix the case when you have only few queries, but super long queries.  Or truncate the query text as in the per database query list.

And what about adding a "queryid != :query"?

powa/query.py
```diff
@@ -122,25 +122,25 @@ def bps(col):
...
             # aligned to kernel ticks. As such, we have to clamp values to 100%
             total_time_percent = lambda x: least(100, (x * 100) /
                                                  total_time)
             cols.extend([
-                kc.reads,
-                kc.writes,
-                total_time_percent(kc.user_time * 1000).label("user_time"),
-                total_time_percent(kc.system_time * 1000).label("system_time"),
+                sum(kc.reads).label("reads"),
+                sum(kc.writes).label("write"),
```
anayrat: shouldn't be `label("writeS")` ?

rjuju: oops, indeed :)  good catch!  fixed and forced-pushed

anayrat: Thanks

powa/qual.py
rdunklau: I'll do that

powa/qual.py
rdunklau: I'll check that

rjuju:
@rdunklau any news?  I'd like to merge this as soon as possible.

powa/query.py
```diff
@@ -209,28 +209,32 @@ def per_sec(col):
...
                      else_=disk_hit_ratio).label("disk_hit_ratio"),
                 case([(total_blocks == 0, 0)],
                      else_=sys_hitratio).label("sys_hit_ratio")])
             from_clause = from_clause.join(
                 kcache_query,
-                kcache_query.c.ts == c.ts)
+                and_(kcache_query.c.ts == c.ts,
+                     kcache_query.c.queryid == c.queryid,
+                     kcache_query.c.userid == c.userid,
+                     kcache_query.c.dbid == c.dbid))
```
rjuju: dbid isn't part of kcache_query.

rjuju:
@rdunklau I pushed two commits to fix the issues I saw for both per-query page and wizard.  Can you check if that's ok and do additional testing?

rjuju:
Rebased against latest master and merged.

PR #71, opened 2019-02-24
Fix disk hit ratio

anayrat:
This PR fix disk hit ratio chart (missing `* 100` factor)

Before:
image

After:
image


rjuju:
Thanks a lot!  I've been wondering why this seemed so wrong without catching a problem with a quick check.  The error is a little bit embarrassing :)

PR #76, opened 2019-06-13
Release 3.2.1

madtibo:
Hello Powa-team,

It would be nice having a 3.2.1 release version that solve some already corrected problems.
That would help people running Powa in production environment while the version 4 is in beta.

Cordialement,

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,4 +1,4 @@
-3.2.1(WIP):
+3.2.1:
   - Add pg_wait_sampling in the config extension list (Julien Rouhaud, thanks
     to Adrien Nayrat for the report)
   - Don't try to detect if a hypotetical index would be used if no suitable


madtibo:
No hurry on our side.
We hit the SQLAlchemy 1.3+ problem and solved it by downgrading the package to 1.2.19.

Thibaut

madtibo:
Well, not very useful PR since I find no way to ask git to apply the version change on 682d0a78311e9ef44f8f355d0903657b189cac4a

Treat this PR more as an issue :-)

rjuju:
There are quite a lot of other fixes in v4, more important and harder to workaround, which certainly won't be available as part of a v3 compatible release as we won't maintain multiple branches.

I'd personally advise to actually test the beta v4 instead of trying to extend an outdated version.  Hopefully, the nature of data stored in powa should ensure that trashing and replacing the data is not a problem.

madtibo:
Hello Julien,
There is no problem with testing the v4, it is just that it is not a good idea to install a beta version on a production server.
The problem I had encountered has been successfully worked-around. But some other problems could block people using version 3.2.0 that cannot install the beta version in their environment.

Cordialement,

rjuju:
I can understand the argument, however the beta had *much mor*e testing than any of the released version.  We're also talking about the UI (which can be installed on a different server), and SQL or plpgsql code.

About v3, SQLAlchemy compatibility is unfortunately just the tip of the iceberg.  Releasing a version 3.2.1 would IMHO send the wrong message, as we *won't* be maintaining that version anymore.  We're now very close to releasing v4, so such release will have a very limited purpose given the number of complaints about SQLAlchemy we had until now.

madtibo:
Ca marche.

Bon courage pour la dernière ligne droite !

rjuju:
Merci !

PR #78, opened 2019-08-02
Use alias instead of hostname:port

anayrat:
~~Remove unnecessary check for srvid=0. Alias is automatically added during first insert in power-archivist.~~ => Only change the query to use alias instead of hostname:port

### DIFF
--- a/powa/framework.py
+++ b/powa/framework.py
@@ -149,10 +149,11 @@ class BaseHandler(RequestHandler):
             return self.current_connection
         else:
             return self.execute("""
-                                   SELECT hostname || ':' || port
-                                   FROM powa_servers
-                                   WHERE id = %(srvid)s
-                                   """, params={'srvid': int(srvid)}
+                                SELECT COALESCE(alias,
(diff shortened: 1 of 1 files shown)


powa/framework.py
```diff
@@ -145,15 +145,13 @@ def get_databases(self, srvid):
             return self._databases
 
     def deparse_srvid(self, srvid):
-        if (srvid == '0'):
```
rjuju: I'd prefer to keep a specific path for the local server.  It's true that the alias is automatically added, but if a user was to remove this alias, it would display nonsensical value (`:0`) instead of the configured repository server connection information.  We shouldn't make life harder for users who don't want an alias for the local server.

anayrat: I put back the old piece of code. It should be fine now.
PS : Thanks for providing docker image, really useful to test :)

rjuju: Ah, I'm glad those images can help working on the UI!

rjuju:
Merged, thanks a lot!

PR #80, opened 2019-08-14
Make sure messages are dismissable

pgiraud:


rjuju:
Thanks!

pgiraud:
Crap. I forgot to update the `powa.min-all.js` file.

rjuju:
I'm not sure that we need to regenerate it every time, especially if you're about to push other modifications?

PR #81, opened 2019-08-14
Prevent footer to be too wide

pgiraud:
Avoids horizontal scroll bar to appear in most cases

Please note that it can be seen in the configuration page easily. There's still a problem with the horizontal scroll bar in some other pages. I'm currently diggin into it.


rjuju:
Indeed, thanks!

PR #82, opened 2019-08-14
Avoid horizontal scroll when there's no data in graphs

pgiraud:
Capture du 2019-08-14 09-12-41
Capture du 2019-08-14 09-13-03


rjuju:
Nice catch.  Do you want to include a grunt run or do you have other changes in js files planned shortly?

pgiraud:
Commit added for the minified js.

rjuju:
Thanks :)

PR #85, opened 2019-08-16
Make sure graph window xMin/Max are updated

pgiraud:
With the previous commit fixing incorrect behavior with data being empty, I introduced a new bug making it impossible for graphs to update their range after a date range change.

Steps to reproduce :
 - open an overview page,
 - data is loaded for the last hour,
 - change the date times.
-> The time range in the charts don't update.


rjuju:
Thanks!

PR #90, opened 2019-11-27
Rework queries

anayrat:
This PR bring following changes:

- Change the predicate order:

`WHERE tstzrange(:from, :to, '[]') @> (records).ts`
to
`WHERE (records).ts <@ tstzrange(:from, :to, '[]')`

I think it is more clear, it is like using:
`WHERE col1 = 4`
instead of
`WHERE 4 = col1`

- Remove an unecessary `UNION ALL` and computation with `min_ts` and `max_ts` : this solved display issue with database object tab. I don't know only this tab was impacted.

Before: https://demo-powa.anayrat.info/server/2/overview/?from=2019-11-26T19%3A46%3A50%2B01%3A00&to=2019-11-27T19%3A46%3A50%2B01%3A00

image

After:

https://dev-powa.anayrat.info/server/2/overview/?from=2019-11-26T19%3A45%3A46%2B01%3A00&to=2019-11-27T19%3A45%3A46%2B01%3A00

image

- Comment some parts of the queries

Note there is still an issue with "general overview":
image

When you zoom on it:
image

### DIFF
--- a/powa/sql/views.py
+++ b/powa/sql/views.py
@@ -58,11 +58,11 @@ def powa_base_statdata_detailed_db():
       )
       AND psh.srvid = :server
     ) AS unnested
-    WHERE tstzrange(:from, :to, '[]') @> (records).ts
+    WHERE (records).ts <@ tstzrange(:from, :to, '[]')
     UNION ALL
     SELECT psc.dbid, psc.userid, psc.queryid,(psc.record).*
     FROM powa_statements_history_current psc
-    WHERE tstzrange(:from,:to,'[]') @> (record).ts
+    WHERE (record).ts <@ tstzrange(:from,:to,'[]')
     AND psc.dbid = powa_databases.oid
     AND psc.queryid IN (
       SELECT powa_statements.queryid
@@ -80,9 +80,7 @@ def powa_base_statdata_db():
  FROM
  powa_databases d LEFT JOIN
  (
-   SELECT srvid, dbid,
-     min(lower(coalesce_range)) AS min_ts,
-     max(upper(coalesce_range)) AS max_ts
+   SELECT srvid, dbid
    FROM powa_statements_history_db dbh
    WHERE coalesce_range && tstzrange(:from, :to, '[]')
    AND dbh.srvid = :server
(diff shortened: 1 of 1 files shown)


powa/sql/views.py
```diff
@@ -49,6 +49,9 @@ def powa_base_statdata_detailed_db():
       SELECT psh.dbid, psh.userid, psh.queryid, psh.coalesce_range,
         unnest(records) AS records
       FROM powa_statements_history psh
+
+      -- coalesce_range can contains more records than we want
```
rjuju: s/contains/contain/?

Maybe also quickly explain why there can be more record than needed, and state the we *first* fetch records overlapping..., which implies that there's at least a 2nd step.

rjuju:
For the remaining issue, did you had a chance to look at the data, to see if the query is returning wrong data or (unlikely but still) the further processing?

powa/sql/views.py
anayrat: Fixed for s/contains/contain/

I agree we should give more explanation for overlapping. I wonder if we should put explanation here : https://github.com/powa-team/powa/blob/master/docs/components/powa-archivist/development.rst

We often use the same pattern:
```
LATERAL
(
  SELECT unnested.dbid, unnested.userid, unnested.queryid,
    (unnested.records).*
  FROM (
    SELECT psh.dbid, psh.userid, psh.queryid, psh.coalesce_range,
      unnest(records) AS records
    FROM powa_<datasource>_history psh
    
    -- coalesce_range can contain more records than we want
    -- so we fetch records overlapping with the requested range
    WHERE coalesce_range && tstzrange(:from, :to, '[]')
    AND psh.dbid = powa_databases.oid
    [...]
    AND psh.srvid = :server
  ) AS unnested
  -- then we keep only record contained in the range
  WHERE (records).ts <@ tstzrange(:from, :to, '[]')
  UNION ALL
  SELECT psc.dbid, psc.userid, psc.queryid,(psc.record).*
  FROM powa_<datasource>_history_current psc
  -- here we can filter only on requested range
  WHERE (record).ts <@ tstzrange(:from,:to,'[]')
  AND psc.dbid = powa_databases.oid
  [...]
  AND psc.srvid = :server
```

anayrat:
> For the remaining issue, did you had a chance to look at the data, to see if the query is returning wrong data or (unlikely but still) the further processing?

I am sure data are good in history tables. The issue is in the query returning data and if I remember, I think the problem is in the processing.

rjuju:
Hi @anayrat, do you have any news on this?

anayrat:
This PR solve the first issue. As I said (https://github.com/powa-team/powa-web/pull/90#discussion_r352672915) I just wonder if we should put explanation here : https://github.com/powa-team/powa/blob/master/docs/components/powa-archivist/development.rst

For the second issue, I spend some time on it but it is more complicated. I think, it should be addressed in another issue in order to not block this PR.

marco44:
Mmm first objection is against yoda conditions... I don't care, it just felt more natural for me writing it this way, but if I'm the only one I don't care, everything being immutable in SQL :)

"Remove an unecessary UNION ALL and computation with min_ts and max_ts : this solved display issue with database object tab. I don't know only this tab was impacted." Ok it seems to solve the issue, but do you have an explanation why ? It's been a while since I had a look at this query :)

anayrat:
> "Remove an unecessary UNION ALL and computation with min_ts and max_ts : this solved display issue with database object tab. I don't know only this tab was impacted." Ok it seems to solve the issue, but do you have an explanation why ? It's been a while since I had a look at this query :)

Sorry for the delay, it took me a while to dig again :)
So the main issue is in this part:

https://github.com/powa-team/powa-web/pull/90/files#diff-829d1062b24654bedfbb2ff563b6c034L837-L855

Especially in theses clauses: `coalesce_range @> min_ts` and `coalesce_range @> max_ts`. We where looking only the ranges containing the min_ts and max_ts but not between them. So I think we were lucky with other tabs, the range contained both min_ts and max_ts interval, but not with this tab.

anayrat:
So this PR can be merged?

rjuju:
Sorry, I totally forgot about this PR.  Thanks for fixing (and bumping), as this is a highly wanted fix :)

PR #92, opened 2019-12-28
Fix global databases metrics query

CyberDem0n:
The missing join condition resulted in incorrect numbers of queries, runtime, page faults, context switches, and block access per second.

The above numbers were N times bigger, where N is the number of databases in the cluster with some registered activity.


rjuju:
Hi @CyberDem0n 

Good catch!  It seems that https://github.com/powa-team/powa-web/commit/c41a88a00269106f339916cea06d2880c5e9e2f5 missed that one.

PR #93, opened 2020-01-15
Remove useless string formatting

CyberDem0n:
Potentially it could lead to sql injection, although chances that it will be exploited are close to zero due to the url format validation.


rjuju:
Hello,

Oh indeed!  I agree that there's probably no security vulnerability here, but it's a terrible coding practice so this patch is more than welcome.

Thanks a lot!

PR #101, opened 2020-06-12
Fix collector processes widget

anayrat:
The application_name do not begin by `'PoWA collector - main thread ` but `'PoWA collector - worker` : 

https://github.com/powa-team/powa-collector/blob/8bdb65b46a8cbc5a293ce1643e11250e48b7765f/powa_collector/powa_worker.py#L171


anayrat:
Sorry, this PR is wrong.

PR #103, opened 2020-06-30
Fix generated where clause on c.calls.

### LINKED ISSUE #102: Wrong where clause lead to wrong generated SQL
Hello,

In several part we have such where clause:
```python
.where(c.calls is not None)
```

Which was translated in SQL with `WHERE true`.

It seems the fix is replace it by:
```python
.where(c.calls != '0')
```

Which is translated in
```
WHERE anon_1.calls != '0'
```

anayrat:
The goal was to remove zero values.
However "c.calls is not None" was translated to "WHERE true"
and "c.calls != None" to "WHERE c.calls IS NOT NULL".

Fix #102

### DIFF
--- a/powa/database.py
+++ b/powa/database.py
@@ -152,7 +152,7 @@ class DatabaseOverviewMetricGroup(MetricGroupDef):
 
         return (select(cols)
                 .select_from(from_clause)
-                .where(c.calls is not None)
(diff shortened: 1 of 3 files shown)


rjuju:
Thanks!

PR #107, opened 2020-07-01
Fixed links to documentation

brainexe:
Currently all link in powa-web to the documentation are broken and leading in a 404.

*Example:*
https://powa.readthedocs.io/en/latest/stats_extensions/pg_wait_sampling.html

*Should be:*
https://powa.readthedocs.io/en/latest/components/stats_extensions/pg_wait_sampling.html


rjuju:
Oh indeed, thanks a lot for spotting and fixing!

PR #111, opened 2020-07-07
Changed some "number" table columns to "integer"

brainexe:
Just a very small UI change...maybe it's just me, but I prefer to strip the decimal places for plain integers, like amouts/counts.

Example from the demo, where you can see the ".00" in the "#" column:
demo

Would then look like:
after_main

Internally the IntegerCell is exending the NumberCell and just setting decimals to 0


rjuju:
That's a very good point, and I totally agree this is way more readable like this!

rjuju:
thanks a lot!

PR #112, opened 2020-07-07
WIP: Re-implemented multiple scales for 2 y-axis

brainexe:
There are some graphs which are using 2 different units, but using the same scale which is not comparable (on absolute values).
E.g. we have millions of index/sec scans but the ratio is always 0-100% -> ratio is completely hidden (when not selecting the ratio series only)
 
I appended some examples...But I'll add more charts with our live values the next days as there the impact might be more visible. But I'd like to get more feedback if there are more cases to keep in mind. I still set minY to 0 as I think that is looking still better and is more comparable.
 
Before:
query_before
After:
query_after

Second example:
Before:
access_before
After:
access_after

### DIFF
--- a/powa/static/js/config.js
+++ b/powa/static/js/config.js
@@ -19,7 +19,6 @@ require.config({
     underscore: "../bower_components/underscore/underscore",
     backgrid: "../bower_components/backgrid/lib/backgrid",
     spin: "../bower_components/spin.js/spin",
-    twix: "../bower_components/twix/bin/twix",
     "backgrid-filter": "../bower_components/backgrid-filter/backgrid-filter",
     "backgrid-paginator": "../bower_components/backgrid-paginator/backgrid-paginator",
     "backbone-pageable": "../bower_components/backbone-pageable/lib/backbone-pageable",
--- a/powa/static/js/powa/views/ContentView.js
+++ b/powa/static/js/powa/views/ContentView.js
@@ -2,9 +2,8 @@ define([
         'powa/views/WidgetView',
         'powa/models/Content',
         'highlight',
-        'moment',
-        'powa/utils/duration'],
-function(WidgetView, Content, highlight, moment, duration){
+        'moment'],
+function(WidgetView, Content, highlight, moment){
     return WidgetView.extend({
         model: Content,
         typname: "content",
--- a/powa/static/js/powa/views/DashboardView.js
+++ b/powa/static/js/powa/views/DashboardView.js
@@ -3,11 +3,7 @@ define([
     'foundation/foundation.equalizer',
     'backbone',
     'powa/views/WidgetView',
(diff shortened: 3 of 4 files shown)


rjuju:
Wow, thanks a lot for working on this.  The current behavior is clearly not ideal, and it also annoyed me quite frequently.

The first graph changes you sent are way better.

About using 0 as Ymin, did you check how it rendered graphs with series having only very high figures?

brainexe:
I also tried around setting a dynamic yMin (as it was mayyybe planned in the commented out code), but it didn't satisfy me. 

Currently I'd still set the fix 0 as min (as it is right now). Example: If a hit ratio is constantly between 97% and 98% I prefer a more or less straight line in the top of the chart. With dynamic yMin the chart would look quite fluctuating when only showing data between 95% ad 100% (needs a second look to exactly check the labels on the y axis).
Here an example WITH dynamic y-min:
dyn_y_min

BTW: here is another examples from production env which are looking bad...with 3 series each which is completely hidden here.
Bildschirmfoto 2020-07-08 um 07 52 31


But I have another idea in mind to make it easier/quicker to read the data: maybe set the color of the y-axis labels in the color of the series (if there is just one)....I'll play around a bit. or maybe someone else has some idea

rjuju:
Yes, the graph can look more spiky than it may really be, but if you look at the scale it should make it clearer.

I'm just worried that keeping an absolute 0-ymin will tend to soften graphs, which can make harder to visually detect some spikes in certain situations.

Does anyone else have some argument here?  ping @marco44 @pgiraud @anayrat

anayrat:
> I'm just worried that keeping an absolute 0-ymin will tend to soften graphs, which can make harder to visually detect some spikes in certain situations.

I agree. If often use dynamic yMin with grafana or munin.

> BTW: here is another examples from production env which are looking bad...with 3 series each which is completely hidden here.
> Bildschirmfoto 2020-07-08 um 07 52 31
> 
> But I have another idea in mind to make it easier/quicker to read the data: maybe set the color of the y-axis labels in the color of the series (if there is just one)....I'll play around a bit. or maybe someone else has some idea

Maybe you can stick serie name on y-axis side. Like grafana : 
image

brainexe:
imho Rickshaw isn't supporting alignment of the legend or colorizing it even more :(

rjuju:
Oh, that's I guess not really surprising but quite disappointing.  Do you know by any chance if that's an issue fixed on some more recent version, or if we're stuck unless we change this lib?

rjuju:
It took a while but the UI is now rewritten using more modern frameworks.  The y-axis seems now behaving correctly, and the series name sticks on the right side.

All the original complaints are fixed I think so I'm closing this PR.  Thanks again for this PR, even if it eventually needed a full rewrite of the frontend.

PR #117, opened 2020-11-06
Add join clause on srvid when joining with powa_statements.

anayrat:
Several queries are built from powa_get<something>data_detailed_db by
passing srvid. Later we join with powa_statements but join on srvid was
missing which often lead to a seqscan on powa_statements.

Thanks to join on srvid we can use powa_statements PK.


marco44 (approved):

rjuju (approved):

rjuju:
Nice catch, this should bring yet more speedup!

Thanks a lot @anayrat !

PR #125, opened 2021-01-10
Fix

Bobsans:
fix bug in indexes optimizations. source: https://www.gitmemory.com/issue/powa-team/powa-web/123/748977067


rjuju:
Oh, wow I was pretty sure that I pushed that fix.  Thanks a lot for noticing and taking care of that!

PR #142, opened 2021-07-22
Fix copyright year in footer

dreis2211:
Hi,

I noticed that the copyright statement at the bottom is still saying 2020. This PR fixes that by just showing the current year, which has the benefit of no further work in upcoming years.

Cheers,
Christoph


rjuju (approved):

rjuju:
I was afraid that this would lead to some discrepancy with the rest of the copyright notice, but apparently we don't have any apart from this one :)

Thanks a lot!

PR #162, opened 2022-11-10
Use vuetify for html in .py

MarionGiusti:


rjuju:
Hi,

Thanks a lot, that's highly appreciated!  However, would it be possible to either squash some commits together if they don't really need to be separated (I see for instance a commit to upgrade vue 2 to 3, and then downgrade it again, and I'm assuming some previous commit introduced it in version 2, some renaming or other mechanical changes can also maybe be batched together and so on), or split the PR in smaller and self-contained batches?

Also, AFAICS there's absolutely no documentation on any change, new tool/dependency introduced or descriptive commit messages.  Looking at the first commit, I only see the commit message "Use vite to build js" with a bunch of changes that I'm not comfortable with since I'm not a web developer.  I really have no idea how I can test the commit, or what that commit is really supposed to do.  I understand that the project was already under documented on that side, but we should use a major rewrite as an opportunity to improve that part rather than keep digging.

PR #163, opened 2022-11-19
wizard: effectively merge nodes with same qualid

pgiraud:
Previous change 291a34d2ecbd67d4e40e1055b873f6f3b9ba8325 was not doing anything actually.

### DIFF
--- a/powa/static/js/powa/models/Wizard.js
+++ b/powa/static/js/powa/models/Wizard.js
@@ -399,9 +399,9 @@ define(['backbone', 'powa/models/DataSourceCollection', 'jquery',
 
           // If we already found a node with the same qualid, we simply need to
           // merge it with the new nodes, otherwise add it to the list.
-          existing = nodes.get(qual.id);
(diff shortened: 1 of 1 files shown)


rjuju:
Hi,

That's quite strange, I'm sure I tested the mentioned commit as I had a usecase that was previously broken and fixed with it.  I'm not really around my computer today unfortunately, so I will double check that tomorrow.

rjuju (approved):

rjuju:
Well, I'm not sure how I ended up with this buggy commit but you're right it was definitely broken so I'm merging this one.

Thanks a lot!

rjuju:
Just for the sake of archive history (I should have put that in the original commit), this part is for merging the same qual on different queries, for instance:

```
SELECT * FROM some_table WHERE id = 1;
SELECT count(*) FROM some_table WHERE id = 1;
```

and have the wizard know that this single qual is related to both queries, and triies to validate the possible index for each.

PR #166, opened 2023-04-24
fixup! Remove sqlalchemy dependency.

pgiraud:
I added a fixup commit for the removal of SQLAlchemy dependency.

### DIFF
--- a/powa/framework.py
+++ b/powa/framework.py
@@ -260,10 +260,11 @@ class BaseHandler(RequestHandler):
             return self.execute("""
                                 SELECT COALESCE(alias,
                                                 hostname || ':' || port)
+                                       AS server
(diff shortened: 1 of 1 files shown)


rjuju:
Ah thanks, I was sure I tested this code path.  Should I merge this PR / rebase / push force or do you want to put this patch at the begining of your big PR?  Getting rid of sqlalchemy wasn't a particularly simple commit so I wouldn't be surprised if you found more similar problems.

pgiraud:
Let's add it directly to the v5 branch. Can you do the merge, then fixup and push force?

rjuju:
Sure!

rjuju:
It should be ok now, thanks for the fix!

PR #181, opened 2023-11-16
Multiple config changes in a single tooltip

pgiraud:
Fix for https://github.com/powa-team/powa-web/pull/165#issuecomment-1813729935


pgiraud:
Screenshot from 2023-11-16 11-16-49

rjuju:
On Thu, 16 Nov 2023, 18:15 Pierre GIRAUD, ***@***.***> wrote:

> Fix for #165 (comment)
> <https://github.com/powa-team/powa-web/pull/165#issuecomment-1813729935>
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/powa-team/powa-web/pull/181
> Commit Summary
>
>    - 9219f31
>    <https://github.com/powa-team/powa-web/pull/181/commits/9219f31a3230624fc28671b8b19b997f792a31e1>
>    Explicitely load Roboto font
>
>
Just a quick note on that commit. For now powa-web is entirely self
contained, in the sense that it won't open any external connection, except
of course the remote servers in some pages (which can entirely be opt out).
I think that a lot of people expect this behaviour and wouldn't be happy
with such a change. I think that at least pgadmin faced quite some backlash
recently with something similar for instance. Could you instead add the
underlying file(s) locally?

pirlgon:
In your second commit, you can specify that only the first five changes are displayed.

powa/static/js/components/dynamic/Graph.vue
```diff
@@ -86,7 +86,7 @@
             changesTooltip.y
           }px) translateX(${tooltipTranslateX(
             changesTooltip
-          )}) translateY(-50%)`"
+          )}) translateY(${tooltipTranslateY(changesTooltip)})`"
```
pirlgon, line 89: The tooltip is now lower and further away from the mouse, is this intentional?

pgiraud: More or less yes. It may look a bit weird at first but I think this is a good workaround for issues with issues I had (z-indexing, or tooltip outside of screen).

pgiraud:
@rjuju I just pushed new commits. The font is now loaded locally.

Also I made some adjustments on the styling. And fixed bugs caught by @pirlgon (list with the first item with no name).

pgiraud:
Here's the new styling which I hope is easier to read.
Screenshot from 2023-11-17 12-00-58

rjuju:
@pgiraud (and @pirlgon) thanks a lot!  It looks quite better than before, I'm verry happy with this change.

I did some other work in the same branch in the meantime, do you plan to rebase the branch or should I cherry-pick the commits?

pgiraud:
Rebased.

powa/static/js/main.js
```diff
@@ -5,7 +5,7 @@ import App from "@/App.vue";
 import dynamicComponents from "@/plugins/powa";
 import VueRouter from "vue-router";
 
-import "@/../styles/main.scss";
+import "@/fonts/Roboto/roboto.css";
```
rjuju, line 8: just to be sure, is the main.scss file still used?  I don't see it referenced anywhere.

pgiraud:
I force pushed new commits but I don't see them here. Last commit should be 8bb862f.

rjuju (approved):

rjuju:
Thanks a lot!  The latest commit is now visible on the UI so it should be ok to merge :)

PR #182, opened 2023-11-16
Fix svg not displayed in Graph title on firefox

pirlgon:
As sugested in https://github.com/vuetifyjs/vuetify/issues/12229#issuecomment-888484775 It's solved in vuetify3.


pirlgon:
image

rjuju (approved):

rjuju:
Ah indeed, good catch!  Thanks a lot for the patch :)

PR #184, opened 2023-12-01
Don't reset handlerConfig when load new dashboard config

pgiraud:
This was leading to a flickr effect on the navigation bar.


rjuju (approved):

rjuju:
I don't really understand what the issue was, but after some testing I don't see anything being worse so I'm fine with it :)

PR #189, opened 2023-12-16
Make sure toprow headers count match columns for grid

pgiraud:
Previously we were expecting the first column not to be provided and to be taken into account in the UI. We now explicitely set it server side.

### DIFF
--- a/powa/database.py
+++ b/powa/database.py
@@ -987,38 +987,30 @@ class DatabaseOverview(DashboardPage):
 
         toprow = [{
                        # query
-                       'merge': True
                    }, {
                        # plan time
-                       'merge': True
                    }, {
                        'name': 'Execution',
-                       'merge': False,
                        'colspan': 3
                    }, {
                        'name': 'Blocks',
-                       'merge': False,
                        'colspan': 4,
                    }, {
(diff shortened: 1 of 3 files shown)


rjuju:
I'm fine with that change.  Looking at the Grid.vue template, I'm wondering if the `merge` attribute is actually still needed?

I see that it was used in the previous js code (powa/static/scss/powa_backgrid.scss) to add some custom css class (https://github.com/powa-team/powa-web/blob/master/powa/static/scss/powa_backgrid.scss#L92-L94) but I don't see anything like that wit the new UI.

pgiraud:
> I'm fine with that change. Looking at the Grid.vue template, I'm wondering if the `merge` attribute is actually still needed?
> 
> I see that it was used in the previous js code (powa/static/scss/powa_backgrid.scss) to add some custom css class (https://github.com/powa-team/powa-web/blob/master/powa/static/scss/powa_backgrid.scss#L92-L94) but I don't see anything like that wit the new UI.

Agreed. It's the right time to remove it. See second commit.

rjuju (approved):
LGTM!

rjuju:
Great, thanks a lot as always, merged!

PR #202, opened 2024-07-08
Misc improvements for the size of built chunks

pgiraud:


rjuju:
looks ok to me, but it would be better to have all the changes in a single commit describing the various approach to reduce the built assets.

pgiraud:
I tend to disagree. Even if most of the commits here help reducing the size of the built assets, each is relevant separately and could possibly be reverted. I prefer we keep them separated.

rjuju:
I don't think that how to best revert each chunk should be what determines a commit boundary, but if you prefer to spam multiple small commits for a single thing go I won't block it.

PR #205, opened 2024-07-24
Make debug mode for run_powa.py optional

pgiraud:


rjuju:
Is it related to https://github.com/powa-team/powa-podman/issues/9?

If yes, I don't really see the point here.  If we don't want to run powa in debug mode we could just run powa-web instead.  The problem is that there still won't be any guarantee that the various compiled assets are up to date, or did I miss something?

pgiraud:
> Is it related to [powa-team/powa-podman#9](https://github.com/powa-team/powa-podman/issues/9)?

It is.

> 
> If yes, I don't really see the point here. If we don't want to run powa in debug mode we could just run powa-web instead. The problem is that there still won't be any guarantee that the various compiled assets are up to date, or did I miss something?

My idea is to run build the assets in the `powa-web-git` image. If not in debug mode, the built assets will be served. If in debug mode, an other container running the ViteJS server will be required.

I'm preparing pull requests in the powa-podman project as well.

rjuju:
Then I still don't think this is the right approach.  IIUC the debug parameter does 2 things:
- print all the executed queries in the log
- use the current "raw" assets rather than the compiled ones

So it looks like this patch would really just be the same as running normal `powa-web`.  If you want to go this route you need to add a new parameter for serving the compiled assets or the raw one that unrelated to the debug mode (although the debug mode should default to serving the raw assets, or more precisely not serving them as apparently in the JS world we need a dedicated daemon running to read a file)

pgiraud:
Thanks for the clarification. I will give this a try with a different approach then.

rjuju:
one final note, I don't think it's crucial for the "normal" powa-web-git image to emit the query in the logs. it's probably way more useful as a powa-web developer, and in this case you should either run it locally or with the dev version of the container. if a user reports a query bug you can just run the same version with one of those images if needed. 

so it might be nice to eventually have that new switch but for now it's probably fine to just rely on powa-web script in powa-web-git. so only compiling the assets during the container build is OK for me. my biggest concern is to make sure that the image size doesn't grow out of control because of it.

PR #209, opened 2024-08-12
Remove dead python code

pgiraud:


rjuju:
does it actively breaks with python 2? of yes, what is the minimum python version supported?

pgiraud:
My bad, psycopg2 actually dropped support for pyton2.7 (and 3.4 and 3.5) in [2.9](https://github.com/psycopg/psycopg2/releases/tag/2_9) 
It even dropped support for [3.6](https://github.com/psycopg/psycopg2/releases/tag/2.9.9).

Python 2.7 officially reached its end of life (EOL) on January 1, 2020. I'd be in favor of dropping its support in PoWA-web as well and follow psycopg2.

Python 3.7 seem to be a reasonable minimum version. What do you think?

rjuju:
I don't think we use anything in psycopg2 that wasn't already there a decade ago.

I know that python 2 is EOL, but the code you want to drop didn't need any maintenance at all, so it seems that the only thing it will achieve is possibly drop support for people who might still use it for no benefit at all.  I don't think we should drop support for anything without a good reason (for instance as I did for dropping support for pg 9.4 in powa-archivist 5)

pgiraud:
I removed the first commit. And only kept the commit in which I drop unused code.

rjuju:
Thanks.  I'd prefer to have both commit merged in a single one as it's the same thing (removing dead code), but with extra details about them, ie. which is the commit that removed the last use of the now dead code.  That can useful in anyone is curious about why those were at some point needed.  The PR should also probably be renamed for the archives's sake.

pgiraud:
Commit squashed and amended with the info I managed to find.

rjuju:
Thanks.  I came to the same conclusion about `urlencode`, but for hybridmethod as far as I can see it was used very early on and became dead code as of 5e7791989b81706cb72d038afff0aee83dfb7eef.

While at it, the comment for `to_json()` in **powa/dashboards.py** is also outdated and could be removed.

rjuju (approved):
LGTM, thanks!

PR #210, opened 2024-08-13
Check python syntax with ruff

pgiraud:


pgiraud:
An example of the github action workflow run: https://github.com/pgiraud/powa-web/actions/runs/10367592193

rjuju:
That's nice!

I used to use pylint (and I updated the code over time to fix pylint warnings for stuff I was working on), how does ruff compare to it?  It seems that the main advantage of rust is to be fast, but we don't have a lot of python code so that shouldn't much of a problem.  On the other hand, ruff seems to handle less checks than pylint, so would it be better to rely on pylint instead, or both?

pgiraud:
The advantages of Ruff is that it's really fast, and it can include rules from differents linters/formatters. We can activate the Pylint linter. I gave it a try (by adding `"PL"`  to the list of select in `.ruff.toml`) and got 45 new errors most of which will be difficult to fix.

Also, Ruff quickly became very popular among python developpers. It already has ~30k stars on github. 

Note : I split into several commits in order to avoid a big bang. I'm not really sure it's better.

.github/workflows/python_lint.yml
```diff
@@ -0,0 +1,29 @@
+name: Check python syntax
+
+on:
+  push:
```
rjuju, line 4: shouldn't it be done on a PR instead?

once you get a failure you can always check locally before pushing again.

.github/workflows/python_lint.yml
```diff
@@ -0,0 +1,29 @@
...
+
+jobs:
+  ruff:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.7", "3.9", "3.11"]
```
rjuju, line 11: is there any value testing multiple version rather than the latest or something? otherwise isn't it possible to have different results on different versions which could lead to the impossibility to have a successful run?

.ruff.toml
```diff
@@ -0,0 +1,3 @@
+[lint]
+# Add "Q" to the list of enabled codes.
+extend-select = ["E4", "E7", "E9", "F", "Q", "I"]
```
rjuju, line 3: some comments explaining what what those codes and why it's a good idea to enable them would be nice

powa/__init__.py
```diff
@@ -1,37 +1,45 @@
 from __future__ import print_function
+
 """
 Powa main application.
 """
-import os
-import re
+import os  # noqa: E402
```
rjuju, line 6: can't we move the from future or some other code around to fix the warning rather than tagging dozens of lines?

.editorconfig
```diff
@@ -8,6 +8,8 @@ insert_final_newline = true
 charset = utf-8
 indent_style = space
 indent_size = 4
+# Default ruff value
+max_line_length = 88
```
rjuju, line 12: that's quite random. most of the code has been written with a line length of 79 characters, and after a quick check it seems that the general rule is still to follow pep 8 which says 79 characters. can we use that and configure ruff to follow it? hopefully it will also reduce the amount of noise in the necessary changes.

powa/collector.py
```diff
@@ -2,8 +2,11 @@
 Dashboard for the powa-collector summary page, and other infrastructure for the
 collector handling.
 """
+
```
rjuju, line 5: is it possible to configure ruff to avoid such annoying changes?

rjuju (changes requested):

rjuju:
I don't think that the number of github star mean anything. otherwise you should be working on mongodb since it has twice the number of stars that postgres has :)

joke aside, I think that a single commit should be better. and then we can add a git blame ignore revs file with its reference to avoid wasting time on it when looking at a file history (there are probably be other commit worth adding to it)

.github/workflows/python_lint.yml
pgiraud: Good catch. I initially chose "push" in order to test this action without opening a PR.

Changed it to "request".

.ruff.toml
pgiraud: I added some documentation comments.

powa/__init__.py
pgiraud: Unfortunately, because of the code computing the `__VERSION__` and used in other files, I couldn't simplify this.

.editorconfig
pgiraud: I changed it back to 79. It unfortunately doesn't reduce the amount of changes a lot.

powa/collector.py
pgiraud: I'm not sure how to change this specific rule. And to be honest, I'm used to it.

.github/workflows/python_lint.yml
pgiraud: The choice I made initialy was very random. I decided to only keep 3.12.

powa/__init__.py
rjuju: ah, that's unfortunate. is there a way to disable the rule for the whole file rather than line by line? that would still be easier

.git-blame-ignore-revs
```diff
@@ -0,0 +1 @@
+046f9e2f4a3db836ddf0b3b7aa9497fca2231af7 # Lint and format with Ruff
```
rjuju, line 1: I will take care of that after the patch is merged. otherwise it's too likely that some stuff will be changed and the commit hash will be outdated and forgotten

powa/collector.py
rjuju: really? I don't understand the rule at all. 

for instance can you explain how to predict all the newlines on top of dashboard.py that look random?

powa/__init__.py
pgiraud: After digging a little bit more, I found a way to ignore some rules at the a file level by adding new lines to the `.ruff.toml` configuration file.

powa/collector.py
pgiraud: `isort` sorts the imports alphabetically and groups them by sections.

https://pycqa.github.io/isort/docs/configuration/custom_sections_and_ordering.html

Here are the default sections: FUTURE, STDLIB, THIRDPARTY, FIRSTPARTY, LOCALFOLDER

rjuju: is it possible to only have 2 groups, one for the locally defined stuff and one for everything else? or just one group for everything.

pgiraud: If this is really important for you, yes, it's possible to sort imports in a single section with the following configuration in `.ruff.toml`.

```
[lint.isort]
no-sections = true
```

I personaly really like the way isort sorts imports and I'm really used to it. By configuring your favorite editor to run lint and format on file save, you won't even have to think about doing the sorting yourself.

If you're really attached to this, I can do the change. Please tell me.

rjuju: I did configure my favorite editor to run a linter and **tell me* about it.  Automatically reformatting on save is annoying because it messes things up like undo.  Besides, if I did this it would mean using the same linter everywhere, and in that case that would be pyflakes and pep8, which I already use everywhere, rather than ruff.

But that misses the main point.  I thought that a linter should help you avoid bug or bad programming patterns.  How exactly does sorting imports in N groups decided by some obscure rules achieve any of that?  What if tomorrow ruff's author decide that a new group would look good, or ninjalint appears and tells you that now imports are better sorted by whatever new rule?  Our local import starts with "powa" so they will be together anyway.  Most of the changes here as far as I can see are similar, they don't really bring much value and mostly looks like "I like it when there is a new line in that case".

pgiraud: I don't think isort rules are "obscure". I like having a clear separation between what's from standard lib, third or first party or known modules.

But OK, I can live with a "no section" sorting. PR updated.

CONTRIBUTING.md
```diff
@@ -15,6 +15,24 @@ This project relies on a few frameworks / libraries including:
 For the following steps, we assume you use PoWA web in debug mode (for example
 by running `./run-powa.py` or using podman dev environment).
 
+### Python syntax
+
+Python syntax must conform to ruff. CI checks new code with ruff.
```
rjuju, line 20: that's not factually correct. ruff doesn't check the syntax (or not only) but also some coding style that have nothing to do with the syntax

powa/collector.py
rjuju: it's still unclear to me what exactly is first party (and since someone opened an issue and they added it to their faq section, it seems I'm not the only one to be confused). it's especially unclear how first party is different from local folder. anyway I'm not really interested in that topic so just removing the section thing works for me.

powa/collector.py
```diff
@@ -2,7 +2,9 @@
...
 from __future__ import absolute_import
+
```
rjuju, line 7: I'm confused. if there's no section anymore, why is a section created there?

CONTRIBUTING.md
pgiraud: Changed.

powa/collector.py
pgiraud: This is because imports from the future library must be done first.

rjuju: ah I see, I guess the idea is that if something is not in alphabetical order it has to be in a different group.  That's probably less surprising generally so fine by me.

Thanks for the clarification.

rjuju (approved):
LGTM, thanks a lot for taking care of it!

PR #222, opened 2024-11-12
Complement information about built assets in RELEASE.md

pgiraud:


pgiraud:
By the way, it would be nice IMO, if building the assets and the building / publishing of the package could be automatically done via the CI when a tag is added.

rjuju:
I'm not sure how you could build the assets when a tag is added as the assets needs to be included in the tree?

RELEASE.md
```diff
@@ -6,7 +6,7 @@
   - [ ] Run `npm ci` to install packages.
   - [ ] Run `npm run build` to build assets (JS and CSS).
 - [ ] Update [the changelog](https://github.com/powa-team/powa-web/blob/master/CHANGELOG) if not already up-to-date.
-- [ ] Commit the changes (assets and changelog).
+- [ ] Commit the changes (new assets files in `powa/static/dist` and changelog).
```
rjuju, line 9: just to be sure, should powa/static/dist/.vite/manifest.json be included?  It's not an asset but it's in that directory.

pgiraud: Yes all files in `powa/static/dist` should be committed. Description changed.

rjuju (approved):

Issue #218, opened 2024-11-12
5.0.0 build error


devrimgunduz:
Hi,

I'm getting the following error while building 5.0.0:

>     File "/usr/lib/python3.13/site-packages/setuptools/_vendor/packaging/version.py", line 198, in __init__
>      raise InvalidVersion(f"Invalid version: '{version}'")
>  setuptools.extern.packaging.version.InvalidVersion: Invalid version: '"5.0.0"'

This happens both on wheel builds and setup.py builds.

Any ideas why?

Thanks!

Cheers, Devrim

rjuju:
hi,

yes I faced the same error when uploading to pypi. apparently the version has to be declared in a single quoted string otherwise you hit this bug, or maybe there's a less lame answer.

but even doing so you still end up with a non functional powa-web as the needed compiled assets are not included in the git, as this part wasn't adequately documented.

I pinged @pgiraud last Sunday and waiting for fixes before I can release a hopefully working 5.0.1.

rjuju:
Hi,

Version 5.0.1 was just released, including the fix for this problem.

PR #223, opened 2024-11-12
Fix version error message when building package

pgiraud:
After the commit ecb8013b5f in which we changed the way the python code is formatted, the "hack" to get the version in setup.py no longer worked.

The current commit fixes this. A better management of the version should be implemented. For example, using setuptools_scm could be a good option.

Related to #218


rjuju (approved):
thanks!

PR #224, opened 2024-11-13
Messages management fixes

pgiraud:


rjuju (approved):
looks good, but could you do a single commit? both are related and needed to fix the same root issue.

PR #227, opened 2024-11-13
Don't fail on SQL format errors

pgiraud:
For example, with the following query the SQL formatter / highlighter raises an error:
``` SQL
SELECT id AS srvid,
                CASE WHEN id = $1 THEN
                   $2
                ELSE
                   COALESCE(alias, hostname || $3 || port)
                END AS alias,
                CASE WHEN id = $4 THEN $5 ELSE hostname END as hostname,
                CASE WHEN id = $6 THEN $7 ELSE port END AS port,
                CASE WHEN id = $8 THEN set.setting
                    ELSE s.version::text
                END AS version
                FROM public.powa_servers s
                LEFT JOIN pg_settings set ON set.name = $9
                    AND s.id = $10
```

Until this can be fixed in the library and in order to prevent PoWA to work improperly (navigation broken), we catch the error and show the unhighlighted value.

### DIFF
--- a/powa/static/js/utils/sql.js
+++ b/powa/static/js/utils/sql.js
@@ -6,6 +6,11 @@ import { formatDialect, postgresql } from "sql-formatter";
 hljs.registerLanguage("sql", pgsql);
 
 export function formatSql(value) {
-  value = formatDialect(value, { dialect: postgresql });
-  return hljs.highlightAuto(value, ["sql"]).value;
+  try {
+    value = formatDialect(value, { dialect: postgresql });
+    value = hljs.highlightAuto(value, ["sql"]).value;
+  } catch (error) {
+    console.error("Could not highlight SQL:", value);
+  }
+  return value;
 }


rjuju (approved):
LGTM.

out of curiosity, what is the error raised by the mentioned query?

pgiraud:
With the above query the error is
```Error: Parse error at token: SET at line 9 column 40
Unexpected RESERVED_CLAUSE token: {"type":"RESERVED_CLAUSE","raw":"set","text":"SET","start":368,"precedingWhitespace":" "}. Instead, I was expecting to see one of the following:
[...]
```

But I've seen other errors (like `Unexpected OPERATOR token: {"type":"OPERATOR","raw":"=","text":"=","start":77}`).

I'll dig this a bit more with small example queries directly against the library. It may have already been fixed.

pgiraud:
For the record, the formatter doesn't like the `set` alias for the `pg_settings` table in the SQL. I reported [an issue](https://github.com/sql-formatter-org/sql-formatter/issues/801) on the `sql-formatter` project.

PR #229, opened 2024-11-14
Change SQL formatter library

pgiraud:
As mention in https://github.com/powa-team/powa-web/pull/227#issuecomment-2475746550, the `sql-formatter` doesn't correctly handles reserved keywords when they are used as alias for tables.

After some searching I found an other library which seem to be more popular and doesn't fail when formatting such queries.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,13 +8,13 @@
       "name": "powa",
       "version": "0.1.0",
       "dependencies": {
+        "@sqltools/formatter": "^1.2.5",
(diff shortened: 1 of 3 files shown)


rjuju (approved):

rjuju:
Apparently this is failing during container image build, see https://github.com/powa-team/powa-podman/actions/runs/11847215631/job/33016547131

pgiraud:
Sorry about that. I'll fix that soon.

PR #246, opened 2024-12-31
Format more files with Ruff

pgiraud:
powa-web and powa.wsgi were not checked by Ruff because the extension for those files don't match default inclusions.

### DIFF
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -1,4 +1,5 @@
 line-length = 79
+extend-include = ["powa-web", "powa/powa.wsgi"]
 
 [lint]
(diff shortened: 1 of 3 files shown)


.ruff.toml
```diff
@@ -1,4 +1,5 @@
 line-length = 79
+extend-include = ["powa-web", "**/*.wsgi"]
```
rjuju, line 2: it's unlikely that new files named *.wsgi will be added but I don't think that there will be any guarantees that they will be written in python, I would just explicitly name it.

rjuju (changes requested):

rjuju (approved):

PR #255, opened 2025-01-22
Fix the sampling of the pgsa query

### LINKED ISSUE #254: Sampling doesn't work on the Global activity/Parallel activity/Backend age views
It comes from this query:
```
FROM 
            (SELECT pgsa_history.srvid,
              dense_rank() OVER (ORDER BY pgsa_history.ts) AS number,
              count(*) OVER () AS total,
              ts,
              datid,
              cur_txid,
              backend_xid,
              backend_xmin,
              backend_start,
              xact_start,
              query_start,
              state,
              leader_pid
              FROM (
                SELECT *
                FROM (
                  SELECT srvid, (unnest(records)).*
                  FROM public.powa_stat_activity_history pgsah
                  WHERE coalesce_range && tstzrange('2025-01-22 08:23:04+0100', '2025-01-22 09:23:04+0100', '[]')
                  AND pgsah.srvid = '1'
                ) AS unnested
                WHERE ts <@ tstzrange('2025-01-22 08:23:04+0100', '2025-01-22 09:23:04+0100', '[]')
                UNION ALL
                SELECT srvid, (record).*
                FROM public.powa_stat_activity_history_current pgsac
                WHERE (pgsac.record).ts <@ tstzrange('2025-01-22 08:23:04+0100', '2025-01-22 09:23:04+0100', '[]')
                AND pgsac.srvid = '1'
              ) AS pgsa_history
              
            ) AS pgsa
            WHERE number % ( int8larger((total)/(100+1),1) ) = 0;
```
The problem is that "total" is not the amount of samples we want to return, it's the total count of records (and we have one record per active session in the snapshot. So If you have 100 sessions, total is off by a factor of 100 for the sampling, and you end up with 1 sample instead of 100

PR incoming

marco44:
It wrongly computes the sampling: the "total" column is not correct in this context: we (most of the time) have more than 1 record per sample. If you have 100 sessions, the sampling will be off by a factor of 100, and you'll end up with only 1 sample

It's a bit ugly to fix, because we need another subquery...

Closes #254

### DIFF
--- a/powa/sql/views_graph.py
+++ b/powa/sql/views_graph.py
@@ -517,9 +517,9 @@ def BASE_QUERY_PGSA_SAMPLE(per_db=False):
 
     # We use dense_rank() as we need ALL the records for a specific ts
     return """
+    (SELECT *,max(number) OVER () AS total FROM
     (SELECT pgsa_history.srvid,
       dense_rank() OVER (ORDER BY pgsa_history.ts) AS number,
-      count(*) OVER () AS total,
       ts,
       datid,
       cur_txid,
@@ -548,7 +548,7 @@ def BASE_QUERY_PGSA_SAMPLE(per_db=False):
         AND (record).backend_type <> 'autovacuum worker'
       ) AS pgsa_history
       {extra}
-    ) AS pgsa
+    ) AS temp) AS pgsa
     WHERE number %% ( int8larger((total)/(%(samples)s+1),1) ) = 0
 """.format(extra=extra)
 
--- a/powa/sql/views_graph.py
+++ b/powa/sql/views_graph.py
@@ -517,38 +517,41 @@ def BASE_QUERY_PGSA_SAMPLE(per_db=False):
 
     # We use dense_rank() as we need ALL the records for a specific ts
     return """
-    (SELECT *,max(number) OVER () AS total FROM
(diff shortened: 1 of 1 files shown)


powa/sql/views_graph.py
```diff
@@ -517,9 +517,9 @@ def BASE_QUERY_PGSA_SAMPLE(per_db=False):
 
     # We use dense_rank() as we need ALL the records for a specific ts
     return """
+    (SELECT *,max(number) OVER () AS total FROM
     (SELECT pgsa_history.srvid,
```
rjuju, line 521: could you indent this as this is now a sub query?

powa/sql/views_graph.py
```diff
@@ -546,7 +546,7 @@ def BASE_QUERY_PGSA_SAMPLE(per_db=False):
         AND pgsac.srvid = %(server)s
       ) AS pgsa_history
       {extra}
-    ) AS pgsa
+    ) AS temp) AS pgsa
```
rjuju, line 549: nit: ideally we should avoid keywords here even if that seems to be allowed. maybe just tmp or something

rjuju (changes requested):

rjuju (approved):
I rebased t he branch and fixed the indentation and query alias

PR #256, opened 2025-01-28
Improve powa_wait_sampling_history query

marco44:
There is a bug: the result is still ok, but it's a performance issue

The left boundary query for powa_base_waitdata_detailed_db on the interval is wrong: the matching interval should be from/from (see other queries that do the same manipulation). So we're doing twice the full interval of the query, once for nothing (and we redo the comparisons on the unnested records on this large one, which takes lots of cpu)


rjuju (approved):

rjuju:
FWIW I had a quick look and didn't see any similar problem in other queries.

@pgiraud could the ruff check shows the diff rather than saying "something would be changed"?  It's not super useful as is :/

rjuju (approved):

PR #258, opened 2025-01-29
Show diff when running ruff format

pgiraud:


pgiraud:
Please review.

rjuju:
I don't see those options in the documentation at https://docs.astral.sh/ruff/formatter/, do you know where they are documented?

What I do see on that page is

> A unified command for both linting and formatting is [planned](https://github.com/astral-sh/ruff/issues/8232).

We should probably wait for that to be merged or specify an explicit version of ruff to avoid breaking all the CI because ruff decided to change again how to format code.  And since it seemed to have happened recently, I would be in favor of fixing the ruff version for as long as we can.

pgiraud:
The `--diff`option has been added to `ruff format` command a while ago. https://github.com/astral-sh/ruff/blob/main/CHANGELOG.md#011

Indeed, the documentation is missing. It's is available via `--help` though.

rjuju:
ok, I would still prefer to hardcode whatever is the current option in the requirements file and make sure everyone and the CI stays on the same version to avoid the same problem as we had last week.

rjuju:
(sorry typo, I meant version not option, updated the previous comment)

rjuju (approved):

PR #259, opened 2025-01-29
Remove not necessary joins on powa_statements

marco44:
After discussion with @rjuju , these IN are not necessary, they may even be harmful.

As they use CPU cycles, let's remove them


rjuju:
agreed.  Those are harmful in the sense that if a new queryid is generated *and* evicted before a snapshot happens, any associated wait even would then be filtered.  This is a really unlikely scenario, and if it does happen the pg_stat_statements overhead is probably so big that just disabling it should double your performance so you won't end up using powa anyway.

rjuju (approved):

PR #262, opened 2025-02-07
Do not display autovacuum workers in the sessions graphs

marco44:
I think it makes sense to keep them only on the vacuum graphs (database objects view)

Without this, we get very long "transactions" and xids displayed in the main graph, and it's confusing

### DIFF
--- a/powa/database.py
+++ b/powa/database.py
@@ -368,7 +368,7 @@ class DatabasePGSAOverview(MetricGroupDef):
 
     name = "pgsa"
     xaxis = "ts"
-    desc = "All backends, including autovacuum workers, are included"
+    desc = "All backends, except autovacuum workers, are included"
     data_url = r"/server/(\d+)/metrics/pgsa_overview/([^\/]+)/"
     backend_xid_age = MetricDef(label="Backend xid age")
     backend_xmin_age = MetricDef(label="Backend xmin age")
--- a/powa/database.py
+++ b/powa/database.py
@@ -1082,7 +1082,7 @@ class DatabaseOverview(DashboardPage):
         db_graphs.append(
             [
                 Graph(
-                    "Global activity (On database %(database)s, except autovacuum)",
+                    "Global activity (On database %(database)s)",
                     metrics=pgsa_metrics[0],
                     renderer="bar",
                     stack=True,
--- a/powa/database.py
+++ b/powa/database.py
@@ -1082,7 +1082,7 @@ class DatabaseOverview(DashboardPage):
         db_graphs.append(
             [
                 Graph(
-                    "Global activity (On database %(database)s)",
+                    "Global activity (On database %(database)s, except autovacuum)",
                     metrics=pgsa_metrics[0],
                     renderer="bar",
                     stack=True,
--- a/powa/database.py
+++ b/powa/database.py
@@ -1084,6 +1084,7 @@ class DatabaseOverview(DashboardPage):
                 Graph(
                     "Global activity (On database %(database)s)",
                     metrics=pgsa_metrics[0],
+                    desc=DatabasePGSAOverview.desc,
                     renderer="bar",
                     stack=True,
                 )
--- a/powa/server.py
+++ b/powa/server.py
@@ -1853,7 +1853,7 @@ class ServerOverview(DashboardPage):
                 Graph(
                     "Global activity (all databases)",
                     metrics=pgsa_metrics[0],
-                    desc=desc=GlobalPGSAMetricGroup.desc,
+                    desc=GlobalPGSAMetricGroup.desc,
                     renderer="bar",
                     stack=True,
                 )
--- a/powa/server.py
+++ b/powa/server.py
@@ -1853,6 +1853,7 @@ class ServerOverview(DashboardPage):
                 Graph(
                     "Global activity (all databases)",
                     metrics=pgsa_metrics[0],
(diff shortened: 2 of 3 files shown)


marco44:
Ok, all the PRs are red because of ruff. @pgiraud do you want me to do a PR to fix this (a ruff format pass…), then we can rebase everything on it and stop having the error ?

powa/sql/views_graph.py
```diff
@@ -546,6 +546,7 @@ def BASE_QUERY_PGSA_SAMPLE(per_db=False):
         AND pgsac.srvid = %(server)s
       ) AS pgsa_history
       {extra}
+      AND backend_type <> 'autovacuum worker'
```
rjuju, line 549: since the graph has associated documentation, the documentation needs to be updated too.  I'm not sure what's the best way to do it.  We could just rename the graph "Non autovacuum backend age" but that sucks.

Ideally we should first introduce a feature to add a "global documentation part" in the help tooltip on top of the existing per-metric documentation, with the detail of what *backends* here exactly means.  That also would also go in the general "better documentation / discoverability" direction that we were talking about last month.

rjuju (changes requested):

powa/sql/views_graph.py
rjuju: To make progress on that I tried to implement that graph description feature in https://github.com/powa-team/powa-web/pull/264, which includes the description that you can change (and hopefully generally improve).

marco44: Ok. Meanwhile, I fixed a stupid bug I generated with this PR: extra doesn't contain a WHERE when we're on the server level :)

We could call that "user backends" or "client-facing backends" ?

rjuju: I now merged https://github.com/powa-team/powa-web/pull/264 so you can add the proper description for all the graphs that rely on this query.

Also, wouldn't it be better to have the new WHERE clause pushed down at the "io_history" level (so writing it in both parts of the UNION ALL?

marco44: I'll add the definition.
For the UNION ALL, I thought it would be easier to read this way, and I don't think there's a lot to win pushing that down, as there is no way it's going to be using an index. So the record will just be filtered one node up the plan (it's a UNION ALL, so no sorting or anything done apart from concatenating the two sets). If you think it's more readable in the UNION ALL, tell me, I'll fix it

rjuju: Yeah the performance will be the same, it was mostly to keep the "dynamic" part only for what is actually supposed to change depending on the "per_db" parameter.

powa/database.py
```diff
@@ -368,7 +368,7 @@ class DatabasePGSAOverview(MetricGroupDef):
 
     name = "pgsa"
     xaxis = "ts"
-    desc = "All backends, including autovacuum workers, are included"
+    desc = "All backends, except autovacuum workers, are included"
```
rjuju, line 371: The "global activity" graph is also impacted?

That also makes me think that it could be nice to have another graph with the number of autovacuum workers somewhere, maybe in the "Database objects" tab (which could also be added in the per-database view).

rjuju (changes requested):

powa/database.py
marco44: Yeah, right, fixed

powa/sql/views_graph.py
marco44: So, fix or no fix ? :)

rjuju: I think it would be better but if you disagree I can live with it :)

marco44: No problem. I'll do it tomorrow though, I'd rather test it

rjuju: that's a good idea:)

marco44: Done

powa/database.py
rjuju: thanks, the one on the per-server indeed needed to be changed, but I actually meant the other graph relying on this datasource that now gets a different behavior, and is called "global activity".  For instance in server.py around line 1853:

```
                Graph(
                    "Global activity (all databases)",
                    metrics=pgsa_metrics[0],
                    renderer="bar",
                    stack=True,
                )
```

This one used to show autovacuum workers right?

marco44: Fixed too, sorry for the delay

rjuju: Thanks, but you're modifying the graph titles, which are now a bit too long I think.

I added a `desc` parameter on the `Graph()` function so that we don't have to do that, can you instead do like the call in line 1103 and add the information about what backends are considered in a new desc parameter, reusing the one set in `DatabasePGSAOverview.desc` and `GlobalPGSAMetricGroup.desc`.

rjuju (changes requested):

marco44:
Fixed, I think
I hope it's ok now

powa/server.py
```diff
@@ -1853,6 +1853,7 @@ def dashboard(self):
                 Graph(
                     "Global activity (all databases)",
                     metrics=pgsa_metrics[0],
+                    desc=desc=GlobalPGSAMetricGroup.desc,
```
pgiraud, line 1856:
```python
                   desc=GlobalPGSAMetricGroup.desc,
``` 
I know, Monday morning is hard.

rjuju (approved):

rjuju:
Thanks, I also tested it locally and it all looks good.

PR #286, opened 2025-05-23
Show cursor position on graphs

### LINKED ISSUE #275: possibility to have some sort of synchronized cursor on all graphs when moving the move over one ?
This is one of the nice features of grafana, would it be possible to have it too ?

pgiraud:
And show the same line (sync) on other graphs on the same page.

Screenshot from 2025-05-23 14-26-31

Fixes #275

### DIFF
--- a/powa/static/js/components/dynamic/Graph.vue
+++ b/powa/static/js/components/dynamic/Graph.vue
@@ -203,7 +203,7 @@ import { useDateRangeStore } from "@/stores/dateRange.js";
 import { useDashboardStore } from "@/stores/dashboard.js";
 import { useDataLoader } from "@/composables/DataLoaderService.js";
 
-const { changes } = storeToRefs(useDashboardStore());
+const { changes, cursorPosition } = storeToRefs(useDashboardStore());
 
 const props = defineProps({
   config: {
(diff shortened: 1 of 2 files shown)


powa/static/js/components/dynamic/Graph.vue
```diff
@@ -875,6 +898,13 @@ svg.chart {
   stroke-width: 1px;
 }
 
+.cursor-line line {
+  stroke: rgb(var(--v-theme-axisgridlinestroke));
```
rjuju, line 902: could you make it a bit more visible?  I think that plain black would work better.

I don't know about others but each time I used a dashboard with this kind of feature the vertical line was quite visible, and I think it makes it easier to locate what you want more quickly.

rjuju (changes requested):
I can't really comment on the code, but I think making the line more visible would help.

powa/static/js/components/dynamic/Graph.vue
pgiraud: Screenshot from 2025-05-26 08-28-42
Screenshot from 2025-05-26 08-29-18

PR #294, opened 2025-07-27
Use the newly introduced "clock_ts" column

rjuju:
powa-archivist 3.1.0 added a new "clocl_ts" column in the pg_stat_activity related tables, to compute more precise age metrics.  There is no guarantee that this column is valued if the extension was just upgraded, so fallback on the previous "ts" column when needed.


rjuju:
Goes with https://github.com/powa-team/powa-archivist/pull/122

PR #348, opened 2026-09-08
fix: take database into account when checking extension version

### LINKED ISSUE #243: "Index Suggestions" errors when hypopg is not installed
In the query overview, if hypopg is not installed for the given database, the "Index Suggestions" in "Predicates" tab loads indefinitely . The corresponding XHR request actually returns a 500 error:

/server/1/metrics/database/obvious/query/9191891907659356179/indexes/?from=2024-12-20%2008%3A28%3A07%2B0100&to=2024-12-20%2009%3A28%3A07%2B0100
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/dist-packages/tornado/web.py", line 1711, in _execute
    result = method(*self.path_args, **self.path_kwargs)
  File "/usr/local/src/powa-web-dev/powa/query.py", line 583, in get
    ind.name = self.execute(
  File "/usr/local/src/powa-web-dev/powa/framework.py", line 653, in execute
    raise e
  File "/usr/local/src/powa-web-dev/powa/framework.py", line 641, in execute
    cur.execute(query, params)
  File "/usr/local/src/powa-web-dev/powa/framework.py", line 64, in execute
    query = resolve_nsps(query, self.connection)
  File "/usr/local/src/powa-web-dev/powa/framework.py", line 91, in resolve_nsps
    return query.format(**connection._nsps)
KeyError: 'hypopg'
```

Actually the flash message cannot be displayed since it's a `KeyError` not a `psycopg2.Error`.

Steps to reproduce:
 - run the `powa-dev-demo.yml` podman compose file,
 - drop the extension in obvious database:
   - `podman exec -it powa-dev-remote-primary bash`)
   - `su postgres`,
   - `psql -d obvious`
   - `DROP extension hypopg;`

pgiraud:
This fixes possible false positive error when we check the version for HypoPG. In the "Optimize this database" wizard for example, where we get "no results to fetch" as reasons for hypothetical index creation errors.

The wizard is still not behaving correctly. It should show a "HypoPG is not installed" warning/error.

A possible side effect is that it also fixes a bug in the `Index Suggestions` block in the `Predicates` tab in the `Query Overview` view. The `index` endpoint doesn't complain anymore with a `extension not found error` since it detect the absence early enough.

Fixes #243


powa/framework.py
```diff
@@ -596,7 +596,7 @@ def has_extension_version(
             raise Exception("No version provided!")
 
         remver = self.__get_extension_version(
-            srvid, extname, remote_access=remote_access
+            srvid, extname, remote_access=remote_access, database=database
```
rjuju, line 600: since the behavior is fixed, it would be good to modify the function description to explain what the behavior is wrt. database being provided or not.

Other than that is seems that none of the callers outside of hypopg will have a different right since they don't specify the database right?

rjuju (changes requested):

powa/framework.py
pgiraud: > since the behavior is fixed, it would be good to modify the function description to explain what the behavior is wrt. database being provided or not.

Done.

> 
> Other than that is seems that none of the callers outside of hypopg will have a different right since they don't specify the database right?

Indeed, the only cases where a database is specified when `has_extension_version` is called are for `hypopg`. Do you think this is a problem? Do you imagine a possible improvement?

rjuju: no, I think it's perfect I just wanted to make sure there wasn't any obscure problem with that as I only have fuzzy memory about this.

thanks a lot!

rjuju (approved):

